PR #47, opened 2020-02-25
convert: Fix converting to nested dynamic map

alisdair:
When converting an object tree to a nested dynamic map (for example, `cty.Map(cty.Map(cty.DynamicPseudoType))` or deeper), it is necessary to unify the types of the child elements of the map.

Previously, we would only apply this process to the innermost leaf object's attributes, which could result in local unification but globally incompatible types. This caused a panic when converting the single top-level object into a map with concrete type, as the types of the sub-trees were incompatible.

This commit addresses this class of bugs in two ways: adding type unification for multi-level map and object structures, and fallback error handling to prevent panics for this class of failure.

The additional type unification process has several steps:

- First, we add a second unify & convert step at the end of the process, immediately before constructing the final map. This second unification ensures that not only sibling elements but also cousins are of the same type.

- We also ensure that generated type information is preserved for empty collections, by returning an empty map of the detected type, instead of the dynamic type. This prevents a multi-step unification bouncing back and forth between concrete and dynamic element types.

- Finally, we add a specific unification procedure for map-to-map conversions. This inspects and unifies the element types for the maps which are being converted. For example, a populated map(string) and an empty map(any) will be unified to a map(string) type.

Tests are extended to cover reduced configurations from the three source Terraform bugs which prompted this work.

## TODO

I'm not satisfied with the error that is printed for Terraform bug 24167:

```
Error: Invalid value for input variable

  on variables.tf line 1:
   1: variable "x" {

The given value is not valid for variable "x": object is required.
```

I don't yet understand why the path to the specific sub-element isn't displayed.

Edit: [this is explained by James in a comment below](https://github.com/zclconf/go-cty/pull/47#issuecomment-594003864). Fixing it in this PR seems out of scope.

### DIFF
--- a/cty/convert/conversion_collection.go
+++ b/cty/convert/conversion_collection.go
@@ -15,18 +15,18 @@ func conversionCollectionToList(ety cty.Type, conv conversion) conversion {
 	return func(val cty.Value, path cty.Path) (cty.Value, error) {
 		elems := make([]cty.Value, 0, val.LengthInt())
 		i := int64(0)
-		path = append(path, nil)
+		elemPath := append(path.Copy(), nil)
 		it := val.ElementIterator()
 		for it.Next() {
 			_, val := it.Element()
 			var err error
 
-			path[len(path)-1] = cty.IndexStep{
+			elemPath[len(elemPath)-1] = cty.IndexStep{
 				Key: cty.NumberIntVal(i),
 			}
 
 			if conv != nil {
-				val, err = conv(val, path)
+				val, err = conv(val, elemPath)
 				if err != nil {
 					return cty.NilVal, err
 				}
@@ -37,6 +37,9 @@ func conversionCollectionToList(ety cty.Type, conv conversion) conversion {
 		}
 
 		if len(elems) == 0 {
+			if ety == cty.DynamicPseudoType {
+				ety = val.Type().ElementType()
+			}
 			return cty.ListValEmpty(ety), nil
 		}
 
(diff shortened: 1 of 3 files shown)


cty/convert/conversion_collection.go
```diff
@@ -15,18 +15,18 @@ func conversionCollectionToList(ety cty.Type, conv conversion) conversion {
 	return func(val cty.Value, path cty.Path) (cty.Value, error) {
 		elems := make([]cty.Value, 0, val.LengthInt())
 		i := int64(0)
-		path = append(path, nil)
+		elemPath := append(path.Copy(), nil)
```
alisdair: These changes throughout are for clarity, now that the `path` parameter can be used multiple times (including in the call to `conversionUnifyCollectionElements`). Always mutating a copy seemed clearer to me than repeatedly mutating the passed parameter.

cty/convert/public_test.go
```diff
@@ -581,6 +591,64 @@ func TestConvert(t *testing.T) {
...
+				}),
+				"b": cty.MapVal(map[string]cty.Value{
+					"x": cty.ListVal([]cty.Value{cty.StringVal("bar")}),
+				}),
+				"c": cty.MapVal(map[string]cty.Value{
+					"x": cty.ListVal([]cty.Value{cty.StringVal("foo"), cty.StringVal("bar")}),
+				}),
+			}),
+		},
+		// reduction of https://github.com/hashicorp/issues/24167
```
mildwonkey:
```suggestion
		// reduction of https://github.com/hashicorp/terraform/issues/24167
```

Possibly my most nitpicky nitpick yet?

cty/convert/unify.go
```diff
@@ -95,6 +100,44 @@ Preferences:
...
+			continue
+		}
+		if unsafe {
+			conversions[i] = GetConversionUnsafe(ty, retTy)
+		} else {
+			conversions[i] = GetConversion(ty, retTy)
+		}
+		if conversions[i] == nil {
+			// Shouldn't be reachable, since we were able to unify
+			return unifyObjectTypesToMap(types, unsafe)
```
alisdair: This looks like a copy & paste error. I don't think this is reachable and it will probably panic if it ever becomes reachable, so I'll try to figure out what the right thing to do here is.

cty/convert/public_test.go
alisdair: Good eye! Thank you, fixed

mildwonkey:
This looks great to me @alisdair! I approve, but please get a second review. (not that I don't trust _your_ work, but mine)

jbardin:
Re: the mismatched type error;  if there is no valid conversion, we've already lost any path information and the `Convert` function only knows that it's not possible, and simplified types are displayed in the message. 

One possibility is making the mismatch errors more detailed and provide the nested type information so the user can at least try to deduce why it failed, but that may be too noisy in the general case. The much larger change would be to thread diagnostics back through `getConversion` to provide precise errors.

PR #56, opened 2020-06-04
convert: Fix panic: heterogeneous tuple with null

alisdair:
Tuples with elements of different types can be converted to homogeneous collections (sets or lists), so long as their elements are unifiable. For example:

```hcl
toset(["a", "b"])     // all elements have the same type
toset(["a", 5])       // "a" and 5 can be unified to string
toset(["a", 5, null]) // null is a valid value for string
```

However, tuples with elements which are not unifiable cannot be converted to homogeneous collections:

```hcl
toset([["a"], "b"])   // no common type for list(string) and string
```

This commit fixes a panic for this failure case, when the tuple contains both non-unifiable types and a null value:

```hcl
toset([["a"], "b", null]) // should not panic
```

The null value was causing the unification process to result in a list or set of dynamic type, which causes the conversion functions to pass through the original value. This meant that in the final conversion step, we would attempt to construct a list or set of different values, which panics.

Fixes hashicorp/terraform#24377

### DIFF
--- a/cty/convert/conversion_collection.go
+++ b/cty/convert/conversion_collection.go
@@ -156,34 +156,45 @@ func conversionCollectionToMap(ety cty.Type, conv conversion) conversion {
 // given tuple type and return a set of the given element type.
 //
 // Will panic if the given tupleType isn't actually a tuple type.
-func conversionTupleToSet(tupleType cty.Type, listEty cty.Type, unsafe bool) conversion {
+func conversionTupleToSet(tupleType cty.Type, setEty cty.Type, unsafe bool) conversion {
 	tupleEtys := tupleType.TupleElementTypes()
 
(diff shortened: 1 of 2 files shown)


cty/convert/conversion_collection.go
```diff
@@ -156,34 +156,45 @@ func conversionCollectionToMap(ety cty.Type, conv conversion) conversion {
 // given tuple type and return a set of the given element type.
 //
 // Will panic if the given tupleType isn't actually a tuple type.
-func conversionTupleToSet(tupleType cty.Type, listEty cty.Type, unsafe bool) conversion {
+func conversionTupleToSet(tupleType cty.Type, setEty cty.Type, unsafe bool) conversion {
```
mildwonkey, line 159: excellent change, tyvm

PR #68, opened 2020-09-24
cty: Add WithMarksDeep

alisdair:
This is to `WithMarks` as `UnmarkDeep` is to `Unmark`. When applying marks to nested values, it can be useful to recursively apply marks to the entire tree. Consider a "tainted" mark applied to a complex object value: carrying this through to every leaf value ensures that no "tainted" values leak as the object is traversed.

CC @pselle

### DIFF
--- a/cty/marks.go
+++ b/cty/marks.go
@@ -364,3 +364,13 @@ func (val Value) WithSameMarks(srcs ...Value) Value {
 		},
 	}
 }
+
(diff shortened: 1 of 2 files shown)


alisdair:
This is unnecessary, because it's how marks already work: https://github.com/zclconf/go-cty/blob/36785d4dc4ac7d38d1a930390a805d23d63a99da/cty/value_ops.go#L708-L712

Whoops!

PR #73, opened 2020-10-22
add Equal method to cty.Type

radeksimko:
As demonstrated by the attached test, this method aids automatic (out-of-the-box) comparison when `cty.Type` needs to be compared for equality, such as in tests.

For context: the [`github.com/google/go-cmp`](https://github.com/google/go-cmp) library always looks for `Equal` method on any type it attempts to compare.

As part of my work on the Terraform language server I keep writing tests which need to compare schemas and these schemas often have `cty.Type` which isn't comparable, so I usually have to explicitly ignore `cty.Type`. This would make tests more accurate and also easier to write (no need for extra options).

### DIFF
--- a/cty/type.go
+++ b/cty/type.go
@@ -39,6 +39,10 @@ func (t Type) Equals(other Type) bool {
 	return t.typeImpl.Equals(other)
 }
 
+func (t Type) Equal(other Type) bool {
+	return t.Equals(other)
+}
+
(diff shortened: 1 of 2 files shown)


radeksimko:
As discussed on Slack this problem can be solved by https://pkg.go.dev/github.com/zclconf/go-cty-debug/ctydebug#pkg-variables so I'm going to close this.

PR #76, opened 2020-11-09
Type.Equals: prevent crash when comparing NilType

radeksimko:
This is useful in combination with https://github.com/zclconf/go-cty-debug and https://github.com/google/go-cmp when comparing structs which may have `cty.Type` field which is optional, i.e. `NilType` is a valid value in that particular context.


apparentlymart (approved):
Seems reasonable! Thanks.

PR #83, opened 2020-12-14
parseint: return a NumberIntVal

mildwonkey:
`parseint()` was returning a `cty.NumberVal` - a float - which caused some odd
behavior when the result, which users expected to be an integer, was used
in other math functions. See the upstream terraform issue
https://github.com/hashicorp/terraform/issues/26365 for some background
context.

The test case added was the example large number that was returning an
unexpected result from parseint. This did change the result of one other test case; it is my belief that this change is correct for the expected (and documented) behavior of `parseint()` but I certainly could be wrong.

### DIFF
--- a/cty/function/stdlib/number.go
+++ b/cty/function/stdlib/number.go
@@ -536,7 +536,7 @@ var ParseIntFunc = function.New(&function.Spec{
 			)
 		}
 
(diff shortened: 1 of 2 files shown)


mildwonkey:
This was the wrong direction, I take it back :)

PR #108, opened 2021-06-10
concat may need to check length with marked vals

jbardin:
If concat is building a tuple type, it will need to check the length of
the arguments which requires unmarking the values.


apparentlymart (approved):
Makes sense to me! Thanks for tracking this down.

PR #110, opened 2021-07-06
flatten: only DynamicVal should cause flatten to become unknown, not a null dynamic value. 

jbardin:
When flattening a list of arguments, the special case of a `null` dynamic value should not force the result to be unknown. We do not check directly for "IsKnown" here, because it is specifically the dynamic type which will cause `flatten` to become unknown, while unknown values alone can be preserved preserved.

https://github.com/hashicorp/terraform/issues/29099


apparentlymart (approved):
Ah, whoops! Thanks!

PR #111, opened 2021-07-06
use big.Float for floor/ceil

jbardin:
Do not convert floor and ceil arguments to float64 internally so as to not lose precision.

This leaves `pow` and `log` as the remaining numeric function which convert to primitive types internally. Both of these are much more complex to implement (which is why they are not included in the `math/big` package itself). The precision offered by `float64` ius likely sufficient for most any log operation, so that probably just leaves `exp` as the last function which needs to be implemented.


apparentlymart (approved):
Looks good to me. Thanks!

PR #112, opened 2021-07-19
cty: handle deep marks in Equals

jbardin:
Equals need to extract all marks from nested values, rather than just
the top level to avoid panicking during comparison.


apparentlymart (approved):
Makes sense to me! Thanks.

PR #114, opened 2021-08-04
cty: AsBigFloat does a shallow copy

jbardin:
Use `Float.Copy` in `AsBigFloat` to ensure that there is no shared data
between the mutable `*big.Float` and the `cty.Value` internal state.

Without the Copy, the given test will output:
```
original value changed from cty.NumberFloatVal(1.9) to cty.NumberIntVal(1)
```


apparentlymart (approved):
Hah, nice catch! I guess callers could (and did) mutate our internal state after all. :upside_down_face: 

Thanks for fixing this.

PR #144, opened 2022-11-07
Strip dynamic types out of null and unknown values recursively

liamcervante:
The original issue is surfaced in terraform as https://github.com/hashicorp/terraform/issues/32109

The root cause is that go-cty is not stripping dynamic information out of null values recursively, it only checks if the value itself is dynamic but not for example is a collection has a dynamic element type.

This approach is the simplest from an architecture point of view, once we decide we are returning a null value then we recurse through the type and replace any dynamic types we would return. However, it's a bit inefficient as it recalculates things like unified types in objects and tuples when this work was already done.

### DIFF
--- a/cty/convert/conversion.go
+++ b/cty/convert/conversion.go
@@ -43,14 +43,14 @@ func getConversion(in cty.Type, out cty.Type, unsafe bool) conversion {
 			out = out.WithoutOptionalAttributesDeep()
 
 			if !isKnown {
-				return cty.UnknownVal(out), nil
+				return cty.UnknownVal(dynamicReplace(in.Type(), out)), nil
 			}
 
 			if isNull {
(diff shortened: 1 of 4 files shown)


cty/convert/conversion.go
```diff
@@ -43,14 +43,14 @@ func getConversion(in cty.Type, out cty.Type, unsafe bool) conversion {
 			out = out.WithoutOptionalAttributesDeep()
 
 			if !isKnown {
-				return cty.UnknownVal(out), nil
+				return cty.UnknownVal(dynamicReplace(in.Type(), out)), nil
```
liamcervante, line 46: @apparentlymart - I specifically wanted to fix Null values with this, but I think it makes sense that unknown and null values behave the same. In case, we don't want that let me know and I'll revert the change to this line.

PR #174, opened 2023-11-01
account for null literal values in function args

jbardin:
The common usage of cty will handle a null literal value as a `NullVal(DynamicPseudoType)`, which is a valid argument to any function parameter allowing nullable values even without AllowDynamicType. Rather than forcing callers to always convert arguments to take this case into account, we return the expected type since the null value is wholly known.


jbardin:
We're going to ensure implementations always use `AllowDynamicType` when dealing with `DynamicPseudoType` here rather than handling another special case to track in the implementation.

Issue #148, opened 2022-12-14
Add support for decoding into structs with a custom tag


gastrodon:
I'd like a feature for a project that I'm working on to be able to do something like `gocty.FromCtyValue`, but for a struct with field tags of a kind other than `cty:""`. Is this within the scope of this project / compatible with design goals?

I've taken a stab at it [here](https://github.com/gastrodon/go-cty) - that branch updates `func structTagIndices` to take a `tag string` instead of always using `cty`, and adds a new function `func FromCtyValueTagged` that takes a new parameter `tag string`

My project uses this function to decode values into structs that are provided by a plugin and everything *seems* to work ( +cty tests pass )

gastrodon:
pr'd #181

PR #181, opened 2024-04-08
Add FromCtyValueTagged

gastrodon:
Like `FromCtyValue`, but you can pass in a different tag to decode with. Update code using literal `"cty"` to use a passed in tag string. Put the insides of `FromCtyValue` into `FromCtyValueTagged`, now `FromCtyValue` calls `FromCtyValueTagged` with literal `"cty"`.


gastrodon:
fixes #148

PR #210, opened 2025-06-05
cty: add `cty.GoIter` for Go 1.23+

giautm:


giautm:
cc @apparentlymart, we may add this as `val.GoIter()` too. On the old version of Go (< 1.23) This function just return a function and doesn't cause any issue.

PR #212, opened 2025-12-24
feat: add support for time.Duration parsing from HCL string

cocktail828:
This change enables parsing time.Duration fields from HCL string values. 

For example:
```go
  type Value struct {
      Dur time.Duration `hcl:"dur"`
  }
```

Can now be defined in HCL as:
```hcl
  data = {
      dur = "10s" // 10h, 20m, 30s, 500ms...
  }
```

### DIFF
--- a/cty/value_init.go
+++ b/cty/value_init.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"math/big"
 	"reflect"
+	"time"
 
 	"github.com/zclconf/go-cty/cty/ctystrings"
 	"github.com/zclconf/go-cty/cty/set"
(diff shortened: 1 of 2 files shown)


apparentlymart:
Hi @cocktail828!

Thanks for working on this, but this sort of magical guessing of what a value was intended to represent doesn't match the abstraction level of this package.

If your motivation for this is using the `gohcl` package in HCL then you could perhaps ask the HCL maintainers if they would be willing to support this shortcut as a helper in _that_ package.

For example, they could potentially implement a rule that it's possible to decode a string value into any type that implements the [`TextUnmarshaler`](https://pkg.go.dev/encoding#TextUnmarshaler) by automatically attempting to use that interface. I believe `time.Duration` implements that interface. I don't know whether the HCL maintainers would accept this as a `gohcl` feature, though.

If the HCL maintainers don't want to support this either then you will need to implement it in your own application by first decoding into a plain string and then calling the duration parser as a separate step. That is the usual way that applications tend to implement this sort of behavior today, in my experience.

Because this functionality is out of scope for this library, I'm going to close this PR.

